PR #5, opened 2019-07-16
Commit for Arrays 1 - Problem 1,2 and 3

Hariae:


jaspindersingh83 (commented):
Good. Use single array to do it now

jaspindersingh83 (commented):
You forgot to add the sample file code. Spiral Matrix looks great.

Hariae:
I want to do it in the way we discussed in class. That way is easier with the thought process, the method which i followed earlier is somewhat confusing at some conditions.

PR #6, opened 2019-07-16
completed  all problems

lavanganji:


Problem1.java
```diff
@@ -0,0 +1,46 @@
...
+        R[length - 1] = 1;  //for the element at index 'length - 1', there are no elements to the right,
+        for (int i = length - 2; i >= 0; i--) {
+            R[i] = nums[i + 1] * R[i + 1];  // calculating the elements from right  to left.
+        }
+
+        // Constructing the answer array
+        for (int i = 0; i < length; i++) {
+            answer[i] = L[i] * R[i];
+        }
+
```
sehaj-grover, line 42: try combining these if possible into a single pass

sehaj-grover (commented):
lavan in diagonal traversal It's not a square matrix so you need m and n both to implement the cases

PR #12, opened 2019-07-16
Please review

AkhilKrishnamurthy:


jaspindersingh83:
Nope we can not use division. In your approach we are using division in second for loop. Rework on problem 1. Submit problem  2 and 3 as well

PR #18, opened 2019-08-11
Array-1 done [Please review]

swaps95shah:


kaurramanpreet (commented):
even if you are able to give brute force solutions for thr problem in first month 
its great !!

PR #19, opened 2019-08-12
Finished Array-1 : Please review

muditabysani:


kaurramanpreet (commented):
attempt the question withour using division as well

PR #23, opened 2019-08-27
Completed Array 1.

priyansh22:
Problem 2 -> Runtime error on LeetCode. Output is correct.


sehaj-grover (commented):
you are missing the empty matrix case in problem 2
check for matrix.size()==0

priyansh22:
I updated the code but it still gives a runtime error on submitting. the output is correct when I run the code.

PR #39, opened 2019-10-22
solutions

Arihant1467:


sehaj-grover (commented):
very good 
Do try the other problem as well

PR #40, opened 2019-10-22
Create Problem-238

pkbaharani:


Problem-238
```diff
@@ -0,0 +1,49 @@
...
+
+
+*/
+class Solution {
+    public int[] productExceptSelf(int[] nums) {
+        
+        int prod=1;
+        int output [] = new int [nums.length];
+        int temp[] = new int [nums.length];
+        int temp1[] = new int [nums.length];
```
sehaj-grover, line 30: output temp1?

sehaj-grover (commented):
very good 
Try the other problems as well

Problem-498
```diff
@@ -0,0 +1,160 @@
+/*
+
+Time Complexity: O(mn) according to me but please help me confirm it.
```
sehaj-grover, line 3: yup we traverse each element once

sehaj-grover (commented):
very good prateek

Problem-238
pkbaharani: I thought of using it, but it was not required

sehaj-grover: Do remove the extra declaration

PR #43, opened 2019-10-22
Array1 Problems completed

TanvirKaur17:


ProductofArrayExceptSelf.py
```diff
@@ -0,0 +1,34 @@
...
+#space complexity = O(n)
+# Leetcode Acceptance = No
+#error in line 17 : index out of range 
+class Solution(object):
+    def productExceptSelf(self, nums):
+        """
+        :type nums: List[int]
+        :rtype: List[int]
+        """
+        res = []*len(nums)
```
sehaj-grover, line 19: fix the list
res = [0 for x in range(len(nums))]
I think it should be like this

sehaj-grover (commented):
please review my comment

neha79091 (commented):
Good. Try to write description in your code.

DiagonalTraverse.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Dec  2 09:15:23 2019
+
+@author: tanvirkaur
+"""
```
neha79091, line 7: Please write complexities.

PR #45, opened 2019-10-23
Arrays-1

Rohanp1196:


sehaj-grover (commented):
very good rohan
try the third problem as well

PR #50, opened 2019-12-02
Update spiral.java

sakshi1705:


spiral.java
```diff
@@ -0,0 +1,45 @@
+/Time Complexity : O(n)
+//Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
```
neha79091, line 3: Complexity is correct if number of elements is n

neha79091 (commented):
Good work .Try other problems as well.

PR #59, opened 2019-12-03
Arrays-1 Please review 

iyersneha:
I had a question regarding the time complexity for the spiral matrix traversal. According to my understanding it should be O(MxN) where M and N are number of rows and columns respectively, and we are going through every element of the matrix once. Leetcode says that the time complexity is O(N) for this. I don't think it is correct. Please let me know which is the correct one. If it is O(N), then why?

### DIFF
--- a/diagonal_traverse.cpp
+++ b/diagonal_traverse.cpp
@@ -1,0 +1,51 @@
+//diagonal traverse 
+class Solution {
+public:
+    vector<int> findDiagonalOrder(vector<vector<int>>& matrix) {
+        //edge case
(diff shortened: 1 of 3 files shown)


jaspindersingh83:
Spiral Matrix in O(MN) yes
Jaspinder Singh

<https://www.linkedin.com/in/jaspindersingh1/>
<https://github.com/jaspindersingh83>

neha79091 (commented):
Good work. You can also mention that you are ignoring the resultant space to calculate space complexity.

PR #117, opened 2020-05-04
Completed

namrataKasar:


Sample.java
```diff
@@ -1,6 +1,169 @@
...
+                }
+            }
+            
+        }
+        
+        return res;
+        
+    }
+
+    // Time Complexity : O(N)
```
sowmyadvn: Please mention what N is here. Because we have rows and columns in matrix, would be O(RC) as we have to iterate through every element. If N is number of elements in matrix, then it is correct.

sowmyadvn:
Solutions are correct Namrata! Please check in-line comments

sowmyadvn:
Solutions are correct Namrata! Please check in-line comments

Sample.java
namrataKasar: Yes N is number of elements. I have updated the time complexity

PR #119, opened 2020-05-04
Solved 3 problems

domzhaomathematics:


array1.py
```diff
@@ -0,0 +1,149 @@
+#Problem 1: Product of Array Except Self
+#Time Complexity: O(n)
+#Space Complexity: O(1) (if we don't count result array)
```
sowmyadvn, line 3: We don't count result space as auxiliary, so you are right here

sowmyadvn:
Good job Dominic! Space/ time complexities are correct

sowmyadvn:
Good job Dominic! Space/ time complexities are correct

PR #138, opened 2020-05-08
Create diagonal.java

tejassjain:


sowmyadvn:
Good job with the problem Tejas! Please try spiral matrix and product except itself problems as well

sowmyadvn:
Good job with the problem Tejas! Please try spiral matrix and product except itself problems as well

PR #139, opened 2020-05-09
Array-1 Completed - Please review

khushal1198:


Array_1_Problem_3.java
```diff
@@ -0,0 +1,135 @@
+//Time Complexity : O(n)
+//Space Complexity: O(n)
+//Run successfully on leetcode
+//Problem : Finding the optimal pattern
+
+//As i was absent in the lecture due to my finals, i tried to come up with my own solution
+//I'm using a linkedhashset to maintain the order and for checking if the element has occureced or not
```
sowmyadvn, line 7: But, what if the follow up question is regarding duplicates?

Array_1_Problem_3.java
```diff
@@ -0,0 +1,135 @@
+//Time Complexity : O(n)
+//Space Complexity: O(n)
```
sowmyadvn, line 2: Space complexity can be improved. Please see the lecture video if you missed the class and try the same problem iteratively and recursively

Array_1_Problem_2.java
```diff
@@ -0,0 +1,81 @@
+//Time Complexity : O(n)
```
sowmyadvn, line 1: For 2D array, it would be clearer if you mention time complexity in terms of rows and columns i.e. O(RC) instead of O(N)

sowmyadvn:
Please review inline comments Khushal. Goob job trying to solve the problem even if you missed the class!

sowmyadvn:
Please review inline comments Khushal. Goob job trying to solve the problem even if you missed the class!

Array_1_Problem_3.java
super30admin: LinkedHashset is an over kill here. Sowmya is right

PR #189, opened 2020-07-21
Completed Arrays-1

alekhya-12:
-Implemented the feedback from previous PRs by giving brute-force solution for problems, where I could.
- For matrix problems, it was a straightforward approach, couldn't think of any other solution.
- Please review. Thanks!

### DIFF
--- a/DiagonalTraversal.java
+++ b/DiagonalTraversal.java
@@ -1,0 +1,63 @@
+/**
+Problem: Diagonal Traversal
+Time Complexity : O(m*n), where m is the number of rows, n is the number of columns of the given matrix.
+Space Complexity : O(1), as we are just returning the result array and not creating an additional space for the purpose of solving.
+(In class, we were told that in most cases, result array that we are asked to return is considered to have a space complexity of O(1)).
+Did it run on leetcode : Yes
+
+Approach: Up and down Directions.
+1. For the purpose of traversing, we make use of directions up and down.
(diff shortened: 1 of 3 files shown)


DiagonalTraversal.java
```diff
@@ -0,0 +1,63 @@
+/**
+Problem: Diagonal Traversal
+Time Complexity : O(m*n), where m is the number of rows, n is the number of columns of the given matrix.
+Space Complexity : O(1), as we are just returning the result array and not creating an additional space for the purpose of solving.
```
samarthjj, line 4: That's right! There wasn't any extra/auxiliary Data Structure created for the purpose of computations!

ProductExceptSelf.java
```diff
@@ -0,0 +1,62 @@
...
+        int[] result = new int[nums.length];
+        int prod = 1;
+        for(int i = 0; i<nums.length; i++) {
+            for(int j = 0; j<nums.length; i++) {
+                if(j!=i) {
+                    prod *= nums[j];
+                }
+            }
+            result[i] = prod;
+        }
```
samarthjj, line 33: Missing return statement! Good work on including both the approaches!

samarthjj:
Great work Alekhya, Please review the inline comments!

samarthjj:
Great work Alekhya, Please review the inline comments!

PR #196, opened 2020-07-21
Completed all 3 problems. Kindly review it

sneharnair:


Problem-1_Product_of_array_except_itself.py
```diff
@@ -0,0 +1,104 @@
...
+#     element.
+# 3. Tke care of the first and last element (don't have left and right neighbors respectively).
+
+
+
+class Solution:
+    def productExceptSelf(self, nums: List[int]) -> List[int]:
+        
+        if nums is None:
+            return -1
```
samarthjj: For all the three approaches, the edge case return type should be an empty array or None, not -1

Problem-3_Spiral_Matrix.py
```diff
@@ -0,0 +1,61 @@
+# APPROACH 1: OPTIMAL SOLUTION
+# Time Complexity : O(n*m), n: number of rows of matrix and m: number of column of the matrix
+# Space Complexity : O(n*m)
```
samarthjj: SC should be O(1)

samarthjj:
Great work Sneha, Please review the inline comments!

samarthjj:
Great work Sneha, Please review the inline comments!

PR #230, opened 2020-09-01
Completed Array-1 problems

sindhusha-parimi:


khushal1198 (commented):
Good work with code and approach!

khushal1198:
Good work with code and approach!

PR #239, opened 2020-09-02
Completed Array 1------Please Review

harinimadhavaram:


Diagonal.java
```diff
@@ -0,0 +1,66 @@
+// Time Complexity : O(N * M) N = rows, M = columns
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Three line explanation of solution in plain english
```
samarthjj, line 4: This is still sort of pseudocode, use sentences that give an overview and are not implementation details. 

An example would be-
# keep a track of three pointers representing row, column and direction. for every
# direction compute three cases which can occur and manipulate the pointers accordingly. 
# repeat the same until the pointers haven't reached the length and width of the matrix

samarthjj, line 4: This is still sort of pseudocode, use sentences that give an overview and are not implementation details. 

An example would be-
-keep a track of three pointers representing row, column and direction. for every
-direction compute three cases which can occur and manipulate the pointers accordingly. 
-repeat the same until the pointers haven't reached the length and width of the matrix

SpiralMatrix.java
```diff
@@ -0,0 +1,65 @@
+// Time Complexity : O(N * M) N = rows, M = columns
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Three line explanation of solution in plain english
+//1. Maintain four pointers Top, Bottom, Left, Right to traverse though the matrix.
```
samarthjj: A good example for this would be- 
-keep a track of four pointers representing each boundary, after traversing and adding
-to array one of 4 side which constitute a spiral traversal, shrink the boundary.
-repeat the same until the boundaries haven't crossed each other.

samarthjj:
Good work Harini, Please review the inline comments on how you can further optimize your description.

samarthjj:
Good work Harini, Please review the inline comments on how you can further optimize your description.

PR #258, opened 2020-09-28
Completed Array 1

Piyush-Kurkure:


Diagonal Traverse.py
```diff
@@ -0,0 +1,20 @@
+# Time Complexity: O(n)
+# Space Complexity: O(n)
+# This code was passsed on LC:  no
+# Problems: Not able to solve
+
```
Lohith-Aswathappa, line 5: The logic seems to be wrong, try out as taught in class and let me know if you still find it difficult.

Lohith-Aswathappa (commented):
Good work, Piyush!
Check inline comment

Lohith-Aswathappa:
Good work, Piyush!
Check inline comment

PR #267, opened 2020-09-29
Completed 1 & 3, need to have a better solution for 2

koshyanand:


Lohith-Aswathappa (commented):
Nice try, Anand!
submit the other problems as well.

Lohith-Aswathappa:
Nice try, Anand!
submit the other problems as well.

PR #273, opened 2020-09-30
Completed Array-1 Sprial Matrix - Please Review

Sneha9p:
Completed Array - 1  Problem 1 and 2 and 3

### DIFF
--- a/Product of Array Except Self
+++ b/Product of Array Except Self
@@ -1,0 +1,23 @@
+Time Complexity - O(N)
+Space Complexity - O(N)
+
+// Multiply the left part and store values and seperately multiyply the right part and store the values
+// When a number is passed and if it sthe 3rd eleemnt left part of the array vlue is multipled with the right part of the array
+
+
+class Solution {
+	public int[] productExceptSelf(int[] nums) {
+		int[] result=new int[nums.length];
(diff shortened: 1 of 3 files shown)


Sprial Matrix
```diff
@@ -0,0 +1,28 @@
+// Time Complexity - O(N)
```
samarthjj, line 1: O(m * n), The entire matrix will have to be traversed at least once.

Sprial Matrix
```diff
@@ -0,0 +1,28 @@
...
+// Space Complexity -  O(N)
```
samarthjj, line 2: O(1), Output DS is not considered as additional space.

Product of Array Except Self
```diff
@@ -0,0 +1,23 @@
+Time Complexity - O(N)
+Space Complexity - O(N)
```
samarthjj, line 2: O(1), Output DS isn't considered for additional space.

samarthjj:
Good work Sneha, Please review the inline comments.
Also please submit .java files and not .txt files.

samarthjj:
Good work Sneha, Please review the inline comments.
Also please submit .java files and not .txt files.

PR #345, opened 2021-01-04
Added all the solutions

saginikhita:


khushal1198 (commented):
Very good work with code and explanation!

khushal1198:
Very good work with code and explanation!

PR #349, opened 2021-01-05
Array-1, completed 1 problem

gangakalburgi:


Lohith-Aswathappa (commented):
Good work, Ganga!
Try the other problem as well

Lohith-Aswathappa:
Good work, Ganga!
Try the other problem as well

PR #401, opened 2021-02-02
Array 1 complete - please review

Sagarudeshi:


Solution1.java
```diff
@@ -0,0 +1,35 @@
+// Time Complexity : O(N)
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode : Y
+// Three line explanation of solution in plain english:
+//find product of elements to left of index,
+//find product of elements to right of index,
+//merge both to find product of all elements except itself
```
yahska-ednihs, line 7: Good Job! with the code and the explanation.

Solution2.java
```diff
@@ -0,0 +1,77 @@
+// Time Complexity : O(MN)
+// Space Complexity : :O(MN)
```
yahska-ednihs, line 2: Good Job! but error in space complexity. Think if you are using any additional space here other than the data structure you get as input or give out as output.

Solution3.java
```diff
@@ -0,0 +1,60 @@
+// Time Complexity : O(N)
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode : Y
+// Three line explanation of solution in plain english
+Iterate on each element in up down fashion. Apply rules when exceeds border
```
yahska-ednihs, line 5: Good Job! But please mention what your symbols stand for. For example, what is N in this case and refer to the comment on the previous problem to correct your space complexity.

PR #407, opened 2021-02-03
Array 1 completed. Please review.

karanm20112:


Problem3.py
```diff
@@ -0,0 +1,34 @@
+# Time Complexity :O(n)
+# Space Complexity :O(n)
+#  Did this code successfully run on Leetcode :yes
+# Any problem you faced while coding this :no
```
yahska-ednihs, line 4: Good Job! Good explanation! Keep up the good work

Problem2.py
```diff
@@ -0,0 +1,38 @@
+# Time Complexity :O(n^m
+# Space Complexity :O(n)
+#  Did this code successfully run on Leetcode :yes
+# Any problem you faced while coding this :no
```
yahska-ednihs, line 4: Good Job! Think on the space complexity. There is a small error. Let us know if you need further explanation.

Problem1.py
```diff
@@ -0,0 +1,63 @@
+# Time Complexity :O(n^m
+# Space Complexity :O(n^m)
+#  Did this code successfully run on Leetcode :yes
+# Any problem you faced while coding this :no
```
yahska-ednihs, line 4: Good Job! Think on the space complexity. There is a small error. Let us know if you need further explanation.

karanm20112: This would be O(1) I guess. Because we are returning that array

Problem2.py
karanm20112: This will be still O(n) right because I am using a map. And approximately it will take O(n) space where n is the number of diagonals.

PR #410, opened 2021-02-04
Create Spiral Matrix

srivastavavertika201:


Spiral Matrix
```diff
@@ -0,0 +1,45 @@
+class Solution {
+    public List<Integer> spiralOrder(int[][] matrix) {
+        if(matrix == null || matrix.length==0 || matrix[0].length ==0){
+            return new ArrayList<>();
```
yahska-ednihs, line 4: Good Job with this problem. Please add other 2 problems as well

PR #422, opened 2021-03-01
Completed Array-1

Nadellka:
Please review the assignment.

### DIFF
--- a/diagonal-order.java
+++ b/diagonal-order.java
@@ -1,0 +1,63 @@
+// Time Complexity :
+o(r*c) where r= rows and c=columns
+// Space Complexity :
+O(1) since we are not using any data structure to store the data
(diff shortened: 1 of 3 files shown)


diagonal-order.java
```diff
@@ -0,0 +1,63 @@
+// Time Complexity :
+o(r*c) where r= rows and c=columns
+// Space Complexity :
+O(1) since we are not using any data structure to store the data
```
nsibal, line 4: I think this is incorrect given that you're using an ArrayList `temp`.

nsibal (commented):
Read inline comments. Otherwise, looks good!

nsibal:
Read inline comments. Otherwise, looks good!

PR #554, opened 2021-06-14
Falling sum, self product, spiral matrix

hharshitarora:


Rooppesh:
Good Job Harshit! Add the TC and SC as well!

Rooppesh:
Good Job Harshit! Add the TC and SC as well!

PR #571, opened 2021-07-05
product of array done

SurbhiZambad2510:


Rooppesh:
Good Job with the solutions! Please add the approach used as well!

Rooppesh:
Good Job with the solutions! Please add the approach used as well!

PR #592, opened 2021-07-07
Please review - Completed Sprint Array-1

Mangesh-Ashtankar:


Rooppesh:
Good Job! Please add the TC, SC and approach as well!

Rooppesh:
Good Job! Please add the TC, SC and approach as well!

PR #637, opened 2021-08-30
Completed 1,2 in Array-1 Please review

sairamya27:


maithrreye:
The code looks good to me but its hardto understand without TC/SCand comment. So please add here after

maithrreye:
The code looks good to me but its hard to understand without TC/SCand comment. So please add here after.

PR #650, opened 2021-09-01
 Self Array and Diagonal Traverese a matrix

danujr:


maithrreye:
the code looks good but still it's hard to decipher without comments explaining approach . Good work on two problem but please finish the other problem . It would be nice to add comments.

maithrreye:
The code looks good but still it's hard to decipher without comments explaining approach . Good work on two problem but please finish the other problem . It would be nice to add comments.

PR #877, opened 2022-04-16
chore: completed with optimal solution 

BitLegion1996:


sanjayvarma199 (commented):
Good Job!! Please include TC and SC

sanjayvarma199:
Good Job!! Please include TC and SC

PR #886, opened 2022-05-10
Array 1

arshmohd:
Implementation of the following problem statement: 
1. Return a result array with product of all elements except ith indexed element itself.
2. Return an array for a given mxn matrix which is a diagonal traversal.
3. Return an array for a given mxn matrix which is spiral traversal.

### DIFF
--- a/Problem-1.java
+++ b/Problem-1.java
@@ -17,27 +17,27 @@ class ProductExceptSelf {
 
     int length = nums.length;
 
-    int[] L = new int[length];
-    int[] R = new int[length];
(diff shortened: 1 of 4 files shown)


Problem-1.java
```diff
@@ -0,0 +1,56 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
shrivatsav, line 2: Please specify what `n` indicates.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

PR #1054, opened 2022-08-03
Please review Array1

NehaGidwani:


NehaGidwani:
Please review

Array1Product.java
```diff
@@ -0,0 +1,26 @@
...
+		int[] result=new int[];
+		int n=nums.length;
+		int rp=1;
+		result[0]=1;
+		for(int i=1;i<=n;i++){
+			rp=rp*nums[i];
+			result[i]=rp;
+		}
+		rp=1;
+		for(int j=2;j>n;j--){ //why is it n-2?
```
jeelpatel18, line 20: Because you are multiplying with its next element, if you try to get element next element for index(n-1) it will give out of bound index exception as array indices are from 0 to n-1 only so.

jeelpatel18: Please re-write again the second for loop because this will give you wrong answer.

jeelpatel18: Did you try to run the code on Leetcode?

jeelpatel18 (commented):
WELL DONE. Review the comments.

jeelpatel18:
WELL DONE. Review the comments.

PR #1123, opened 2022-09-27
Completed

shrutinag04:


diagonalOrder.java
```diff
@@ -0,0 +1,52 @@
+//time complexity: O(n)
```
jeelpatel18, line 1: m*n is the TC

jeelpatel18: SC is constant as you are returning result array

spiral.java
```diff
@@ -0,0 +1,48 @@
+//time complexity: O(n)
+//space complexity: O(n)
```
jeelpatel18, line 2: TC is m*n and SC is O(1)

jeelpatel18 (commented):
Good job.

jeelpatel18:
Good job.

PR #1158, opened 2022-10-24
Arrays-1: Please review

pavansai1:


p1.java
```diff
@@ -0,0 +1,36 @@
+// TC: O(n)
+// SC: O(1)
```
daredevilteja, line 2: SC is O(n)

daredevilteja:
Good Work! Keep it up

PR #1174, opened 2022-11-14
Completed Array1 | please review

Reethuch:


problem2.py
```diff
@@ -0,0 +1,43 @@
+# 498. Diagonal Traverse
+# TC : O(N)
```
daredevilteja, line 2: TC is O(m*n)

daredevilteja:
Good work! Keep it up.

PR #1257, opened 2023-03-30
array-1 done, please review.

UjasShah:


array_diagonal_order.py
```diff
@@ -0,0 +1,45 @@
+# Time Complexity : O(n) because we are traversing the array once
```
osandeepreddy, line 1: TC is O(m*n)

array_spiral_order.py
```diff
@@ -0,0 +1,43 @@
+# Time Complexity : O(n) because we are traversing the matrix once
```
osandeepreddy, line 1: TC is O(m*n)

osandeepreddy:
TC is O(m*n)

PR #1329, opened 2023-07-14
Complete Array-1

PurnaSahithiAdduri:
Complete Array-1

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,28 @@
+// Time Complexity :O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : yes
+// Three line explanation of solution in plain english
+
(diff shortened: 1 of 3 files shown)


Problem1.java
```diff
@@ -0,0 +1,28 @@
+// Time Complexity :O(n)
+// Space Complexity : O(n)
```
ishika2741: Try to submit optimized space solution as well.

Problem2.java
```diff
@@ -0,0 +1,47 @@
+// Time Complexity :O(m*n)
+// Space Complexity : O(n)
```
ishika2741: Space would be O(1) as output space does not count

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1397, opened 2023-10-13
Completed Array - 1

vinUrlam:


diagonal-matrix.py
```diff
@@ -0,0 +1,43 @@
+# Time Complexity : O(MN)
+# Space Complexity : O(M+N)
```
ishika2741, line 2: Try to submit optimised solution as well

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1399, opened 2023-10-14
Added solutions for spiral_order and product_array

nithind-sudo:


diagnoal_traverse.py
```diff
@@ -0,0 +1,39 @@
+# TC : O(rows * cols)
+# SC : O(rows * cols)
```
ishika2741, line 2: Check space complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1426, opened 2023-11-23
Completed Array-1

bhavana-nair:


ishika2741 (commented):
Good work! Kindly add Tc and SC as well.

ishika2741:
Good work! Kindly add Tc and SC as well.

PR #1567, opened 2024-07-06
completed Array-1

dhikki-study:
##########238. Product of Array Except Self######################################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I was unable to find the solution after class understood could be done using pre/post production of elements

// Your code here along with comments explaining your approach in three sentences only
#first tool pre product and than took post product and multiple each element at there position to find the result

        

##########498. Diagonal Traverse###################################################
// Time Complexity : mXn
// Space Complexity : mXn
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : solved after class, so have better idea

// Your code here along with comments explaining your approach in three sentences only
Identified the condition which will reach the matrix and applied specific checks for them

##########54. Spiral Matrix########################################################
// Time Complexity : mXn
// Space Complexity : mXn
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I thought of doing the question using sets but after class solved it using iteration. Also the code was failing when initial while condition was mutated so need to check video to identify the extra condition required

// Your code here along with comments explaining your approach in three sentences only
Found 4 pointers, rowmin, column, rowmax, colmax, once traversed each element is reduced

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,127 @@
+##########238. Product of Array Except Self########################################################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I was unable to find the solution after class understood could be done using pre/post production of elements
+
+// Your code here along with comments explaining your approach in three sentences only
+#first tool pre product and than took post product and multiple each element at there position to find the result
+
+class Solution:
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1619, opened 2024-10-07
Completed

Ameya-20:


Nandha2Tech (commented):
Great Work! Please check the spiral matrix logic. You are alternating between moving left to right and right to left row-wise, which does not cover the full matrix in a spiral pattern.

Nandha2Tech:
Great Work! Please check the spiral matrix logic. You are alternating between moving left to right and right to left row-wise, which does not cover the full matrix in a spiral pattern.

